PR #4, opened 2019-07-20
Completed Solution to Problem 1

gauravj09:


ashmeetdeol19:
Solutions to both the problems are great.
Also Submit Solution to Problem3 as well.

PR #8, opened 2019-07-20
Problem 1 and Problem2 - Please review

hrshah5:


Problem2.java
```diff
@@ -0,0 +1,26 @@
...
+            arr[i]= nums2[j];
+            i++;
+        }
+        Arrays.sort(arr);
+        for(i = 0; i<nums1.length; i++){
+            nums1[i]= arr[i];
+        }
+        
+    }
+}
```
ashmeetdeol19: You should also try the optimal solutions and submit solution for Problem 3 as well.

PR #9, opened 2019-07-20
Completed Two-Pointers-2- Please Review

bharatreddypothireddy:


bharatreddypothireddy:
Completed All Problems.

super30admin:
Great job Bharat
1. Problem 1 solutions is good and your explanation is to the point.
2) Problem 2 solution with LinkedList is good way of thinking about various other data structures but try two pointer solution as well.
3) Problem 3: No need to get confused about negative elements. Just make a nextIndex or getIndex separately which takes care of both positive and negative numbers.

PR #18, opened 2019-09-05
Completed. Kindly review the code.

DeekanshaTandon:


merge_sortedArray.java
```diff
@@ -0,0 +1,47 @@
...
+                j--;
+                i--;
+                k=k-2;
+
+
+            }
+
+
+        }
+        while(i >= 0 && k >= 0){
```
kaurramanpreet: do we need this while loop?

PR #20, opened 2019-09-06
Completed pls review.

PATIL1695:


RemoveDuplicates.py
```diff
@@ -0,0 +1,37 @@
...
+position at 7 - i+1 != i -1 -> count += 4 = 5
+position at 8 - i+1 != i -1 -> count += 5 = 6
+
+           |   | |     | | | |   ----> '|' indicates the elements considered.
+example - [1,1,1,2,2,2,2,3,3,4]
+
+finally take into account the last index also so return count+1 = 7 = [1,1,2,2,3,3,4]
+Basically we are just taking into account the first and last index elements of the repeated sub-arrays in
+the main array.
+"""
```
kaurramanpreet: great approach!!
i really appreciate the way you think!!

jaspindersingh83: This is good approach. I agree with @kaurramanpreet

PR #22, opened 2019-09-07
Please REview

Rakshitha94:
@kaurramanpreet  I did not get the duplicates problem . I want it to be xplained once

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# Default ignored files
+/workspace.xml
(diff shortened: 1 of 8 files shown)


super30admin:
where are you struck in that problem?let me know so that i can explain accordingly

Rakshitha94:
in Merge array problem, why do we have one more while loop outside to append the reamining elements?

we already check that in the first while loop right?
 while (  i >= 0 and j >= 0): 


In duplicates problem: its not executing in leetcode

kaurramanpreet:
your duplicates problem is not executing on leetcode as there is indentation problem in return statement
correct it and it will run successfully

kaurramanpreet:
and for merge array problem
as if i have merged 2 arrays and some numbers are still remaining in the 2nd array then we should be having one more loop to append them as well

PR #23, opened 2019-09-08
Problem 1 and 2 committed. Problem 3 pending

manishjangid:


Problem1.c
```diff
@@ -0,0 +1,171 @@
...
+
+Given nums = [0,0,1,1,1,1,2,3,3],
+
+Your function should return length = 7, with the first seven elements of nums being modified to 0, 0, 1, 1, 2, 3 and 3 respectively.
+
+It doesn't matter what values are set beyond the returned length.
+
+Clarification:
+
+Confused why the returned value is an integer but your answer is an array?
```
super30admin: because the main function is made in such a way that it will iterate the returned number of times over the array which was passed as reference in the function

PR #33, opened 2019-10-26
Create merge_two_sorted_arrays.java

SuperStriker:


search_2D_matrix.java
```diff
@@ -0,0 +1,22 @@
+class Solution {
```
SuperStriker: TC: O(n)
SC:O(1)
Leetcode : Successful
Runtime: 5 ms, faster than 100.00% of Java online submissions for Search a 2D Matrix II.
Memory Usage: 42.7 MB, less than 100.00% of Java online submissions for Search a 2D Matrix II.

Approach:
Start with index last row and first column
if target is > m[i][j] , then move right,if target is < m[i][j] , move left
Repeat till target is found or index i becomes < 0 or j becomes > n-1, which means target is absent in matrix.

PR #41, opened 2019-12-06
Completed Two-Pointers-2

karthikmk:


searchMatrix2.cc
```diff
@@ -0,0 +1,27 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : It ran successfully
+// Any problem you faced while coding this : None
```
neha79091: Time complexity will be O(m+n), where m and n are rows and columns respectively.

mergeSortedArrays.cc
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
neha79091: It will be O(m+n)

PR #44, opened 2019-12-11
Completed two pointer problems

dasgprat:
Please review the PR and also let me know in case of any issue or concerns.

### DIFF
--- a/Problem_36.java
+++ b/Problem_36.java
@@ -1,0 +1,34 @@
+/**
+LeetCode Submitted : YES
+Time Complexity : O(N)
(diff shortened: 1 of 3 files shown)


Problem_36.java
```diff
@@ -0,0 +1,34 @@
+/**
+LeetCode Submitted : YES
+Time Complexity : O(N)
+Space Complexity : O(1)
```
neha79091: Time complexity will be O(m+n).If you are thinking N as M+n ,then it is fine.

PR #46, opened 2019-12-13
Submitting TwoPointers-2

rakhil163:


mergeSorted.java
```diff
@@ -0,0 +1,21 @@
+Time Complexity-O(n)
+Space Complexity-O(1)
```
neha79091: time complexity is O(n+m)

PR #50, opened 2019-12-16
adding solutions for Two pointers array exercise 2

amitkc1:


MergeSortedArray.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
neha79091: Time complexity should be O(m+n)::m and n be the number of first and second array respectively

PR #61, opened 2020-01-11
please review two pointers 2

BhaveshTadikonda:


super30admin:
Good job with both problems Bhavesh

PR #74, opened 2020-02-21
before class brute force solutions

pbandyop:
@Lohith-Aswathappa


Lohith-Aswathappa:
It's a good thought process! Try to implement it as well

pbandyop:
after class solutions (pending from last week) - please review
@Lohith-Aswathappa

PR #77, opened 2020-02-21
Initial update. Will update with binary search sooner

saiswaroopreddy:


Lohith-Aswathappa:
Nice try!
Try to solve the 2nd problem without inbuilt functions
3rd problem returns true even if the element is not present in the matrix

PR #106, opened 2020-04-14
Solution for merge sorted array problems (2 pointers)

vsaylee4:


devashishnyati:
Good job Saylee!
Now try to finish all the pending PR's.

PR #111, opened 2020-05-08
Completed first 2 problems

namrataKasar:


Sample.java
```diff
@@ -1,7 +1,124 @@
...
+                }
+            }
+            j++;
+        }
+        
+        return i;
+        
+    }
+    
+    // Time Complexity : O(N)
```
sowmyadvn, line 43: Two array sizes are M and N, we are traversing through both. So, complexity has to be O(M+N) here

sowmyadvn:
Good job with the solutions! Please check in-line comments Namrata

sowmyadvn:
Good job with the solutions! Please check in-line comments Namrata

Sample.java
namrataKasar: We are traversing both arrays simultaneously, so it should be O(N) where N is length of first array.

sowmyadvn: Then, that is right. I thought N as actual size of first array without considering space for second array's elements.

PR #113, opened 2020-05-08
Update Sample.java. Please review two-pointer-2. 

Abinashi7:


iivaishnavii:
Great job with all the three problems,Abinashi! Neat comments.

PR #123, opened 2020-05-15
Update Sample.java

ashutosh-senapati:


iivaishnavii:
Great job with all the three solutions, Ashutosh..keep it up!

PR #128, opened 2020-05-26
Solved

tejassjain:


search_2D_matrix_ii.java
```diff
@@ -0,0 +1,21 @@
+//Time Complexity: O(m*n)
```
sowmyadvn, line 1: Before while loop terminates, we are incrementing row m times and decrementing column n times. Correct? So, we are not iterating through entire matrix, but m+n iterations. So, time complexity is O(M+N)

sowmyadvn:
Good work Tejas! Please review inline comment regarding time complexity

sowmyadvn:
Good work Tejas! Please review inline comment regarding time complexity

PR #132, opened 2020-06-11
Pleaase review

sumukhballal:


mergeSortedArray.java
```diff
@@ -1,3 +1,10 @@
+// Time Complexity : O(n)
```
AbhiniveshP, line 1: There are 2 arrays and so be specific when you mention O(n)

AbhiniveshP:
Nice work Sumukh !!!
Please be careful while mentioning Time complexities

mergeSortedArray.java
sumukhballal: Yeap got it! Sorry thanks

PR #145, opened 2020-06-11
Two_Pointers_2 completed!

sarthaksugandhi9:


AbhiniveshP:
Nice work Sarthak !!!
What if you are allowed to have k duplicates instead of 1 allowed duplicate?

AbhiniveshP:
Nice work Sarthak !!!
What if you are allowed to have k duplicates instead of 1 allowed duplicate?

sarthaksugandhi9:
I need to change condition <= 2 with <=k. Then the code will be applicable
for any value.

PR #161, opened 2020-06-15
Two-Pointers-2 problems : Please review

nareshkumar-ss:


AbhiniveshP:
It is nice that you have included brute force solutions as well for all the PRs.
Try not to mention time complexities as O(nlogn + n) as both the terms are functions of n, you have to include only the higher-order term which is nlogn in this case.

AbhiniveshP:
It is nice that you have included brute force solutions as well for all the PRs.
Try not to mention time complexities as O(nlogn + n) as both the terms are functions of n, you have to include only the higher-order term which is nlogn in this case.

PR #165, opened 2020-06-15
Two-Pointers-2 completed: please review

AkshayGyara:


AbhiniveshP:
Nice work on all 3 questions !!!
Time Complexity for mergeSortedArrays would be O(m + n) and not O(n) as it is varying in terms of both m and n. You could also say O( max (m, n) ) but not just O(n)

AbhiniveshP:
Nice work on all 3 questions !!!
Time Complexity for mergeSortedArrays would be O(m + n) and not O(n) as it is varying in terms of both m and n. You could also say O( max (m, n) ) but not just O(n)

PR #175, opened 2020-07-24
completed

satishsolanki1990:


samarthjj:
Good work Satish, Please use template from sample.java file. 
Going forward please use complete English sentences for your approach instead of broken sentences or pseudocode.

samarthjj:
Good work Satish, Please use template from sample.java file. 
Going forward please use complete English sentences for your approach instead of broken sentences or pseudocode.

PR #181, opened 2020-07-25
Please Review - Two Pointers -2 

firoz1905:


Sumesh-H:
Good work, Firoz! All TC & SC are correct.

Sumesh-H:
Good work, Firoz! All TC & SC are correct.

PR #197, opened 2020-07-31
Completed, please review

JaneJianLee:


solution2.py
```diff
@@ -0,0 +1,28 @@
+#https://leetcode.com/problems/merge-sorted-array/
+# // Time Complexity : o(n)
```
samarthjj, line 2: O(m+n)

solution1.py
```diff
@@ -0,0 +1,34 @@
+# https://leetcode.com/problems/remove-duplicates-from-sorted-array-ii/
+# // Time Complexity : o(n)
+# // Space Complexity : o(1)
+# // Did this code successfully run on Leetcode :yes
+# // Any problem you faced while coding this : maintaining three pointers
```
samarthjj, line 5: Fast is incremented no matter what, instead of maintaining three pointers you could use a for loop. Bringing down to manual track of two pointers!

samarthjj:
Good work Jane, Please review the inline comments!

samarthjj:
Good work Jane, Please review the inline comments!

PR #227, opened 2020-10-01
Completed

siddhesh25:


Problem3.py
```diff
@@ -0,0 +1,28 @@
+# Time Complexity: O(N)
```
Lohith-Aswathappa, line 1: check time complexity

Lohith-Aswathappa (commented):
Good work, Siddhesh!
Check inline comment

Lohith-Aswathappa:
Good work, Siddhesh!
Check inline comment

PR #228, opened 2020-10-01
init comit for two pointer 2

amitsingh5949:


ramya-sruthi-d (commented):
Good work. Try to put even more detailed comments.

ramya-sruthi-d:
Good work. Try to put even more detailed comments.

PR #231, opened 2020-10-01
two pointers 2

HarshaVardhan-Kaki:


ramya-sruthi-d (commented):
Well done. Try the optimal solution for removing duplicate elements as well. Also, explain your approach.

ramya-sruthi-d:
Well done. Try the optimal solution for removing duplicate elements as well. Also, explain your approach.

PR #369, opened 2021-02-04
Completed Two Pointers 2 - Please review

rachana-code:


kannan-mrugapphan (commented):
Good Job. Also briefly discuss about the approach as well through comments

kannan-mrugapphan:
Good Job. Also briefly discuss about the approach as well through comments

PR #428, opened 2021-04-22
Completed Two-Pointers-2

Nitishmane:


maithrreye:
Good Job Nitish ! Comment on your code and TC/SC would be appreciated !

maithrreye:
Good Job Nitish ! Comment on your code and TC/SC would be appreciated !

PR #443, opened 2021-05-10
Completed Pointers-2- P2,P3

Beryl2208:


Lohith-Aswathappa (commented):
Good work, Beryl!
Try the other problems as well.

Lohith-Aswathappa:
Good work, Beryl!
Try the other problems as well.

PR #460, opened 2021-06-10
added two pointers 2

pandas5308:


maithrreye:
Comment about your approach would be helpful because its hard to decipher your approach ! Add TC and SC .

maithrreye:
Comment about your approach would be helpful because its hard to decipher your approach ! please add TC and SC .

PR #467, opened 2021-06-11
Completed Two pointers Please Review 

vigneshdotkumar:


mergesortedarray.java
```diff
@@ -0,0 +1,30 @@
...
+        while(ptr2>=0)
+        {
+            nums1[ptr3]=nums2[ptr2];
+                ptr3--;
+                ptr2--; 
+        }
+    }
+}
+
+//time complexity : O(m*n)
```
sridhar-anish, line 29: Please rethink the TC of the solution

sridhar-anish (commented):
Good job with everything else

sridhar-anish:
Good job with everything else

PR #505, opened 2021-06-16
Two-pointers

pratikshakapse5:


mergeSorted.cpp
```diff
@@ -2,36 +2,28 @@ class Solution {
...
+            j--;
             k--;
         }
-        while(i>0)
-        {
-            nums1[j]=nums2[i];
-            i--;
-        }
-        
+
```
sridhar-anish, line 27: Good Job, but make sure to include the time and space complexity

sridhar-anish (commented):
Good job, with everything else

sridhar-anish:
Good job, with everything else

PR #520, opened 2021-07-08
Completed Two-Pointers-2 - Please Review

Suramya28:


Suramya28:
Problem1.java

//time: O(n), space: 0(1)
class Solution {
    public int removeDuplicates(int[] nums) {
      
         int res = 2;
        for(int i=2; i<nums.length; i++)
        {
            if(nums[i] != nums[res-2])
            {
                nums[res] = nums[i];
                res++;
            }
        }
        return res;
    }
}

PR #613, opened 2021-10-10
Two-Pointers-2 completed please review

srinathsai:
As i mentioned i messed and submitted two pointers-2 two files in two pointers-1 .so two of two pointer-2 files are in two-pointers-1 and two of two pointers-1 are in here.Sorry for that

### DIFF
--- a/remove duplicates.java
+++ b/remove duplicates.java
@@ -1,0 +1,30 @@
+//timecomplexity :-o(n^2);
+//space complexity:-0(1);
+//did it run on leetcode:- no
+//any problem you faced:- got wrong answers and time limit exceeded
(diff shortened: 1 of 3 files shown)


remove duplicates.java
```diff
@@ -0,0 +1,30 @@
+//timecomplexity :-o(n^2);
+//space complexity:-0(1);
+//did it run on leetcode:- no
+//any problem you faced:- got wrong answers and time limit exceeded
```
manojvarmarudraraju, line 4: write down the flow on a whiteboard, which will give you clarity on how to move the pointers, one pointer pointing the right place the other pointer will be iterating over the array

manojvarmarudraraju (commented):
Good job!!

manojvarmarudraraju:
Good job!!

PR #728, opened 2022-02-18
amitsingh5317 init two pointer 2

amitsingh5317:


_240_Searcha2DMatrixII.java
```diff
@@ -0,0 +1,21 @@
+// space complexity - o(1)
+  //time - o(n+m)
```
shrivatsav: Please specify what ` m and n` indicate.

shrivatsav (commented):
Good Work! Please address the feedback.

shrivatsav:
Good Work! Please address the feedback.

_240_Searcha2DMatrixII.java
amitsingh5317: Thank you for the feedback. I have updated the comments in complexity.

PR #865, opened 2022-06-12
Done with all exercises for two pointers

kedark3:


search2DMatrix2.py
```diff
@@ -0,0 +1,27 @@
+'''
+Approaches:
+1) Linear search - O(m*n) time to loop over the matrix
```
ishika2741, line 3: Kindly check the complexity.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

search2DMatrix2.py
kedark3: Wouldn't that be O(m*n)? Looping nested over m rows and n columns and nested loop meaning multiplication. Please clarify.

PR #891, opened 2022-06-25
Added files

anoopkak:


jeelpatel18 (commented):
Good work. Please add TC and SC and also submit the 3rd problem as well .

jeelpatel18:
Good work. Please add TC and SC and also submit the 3rd problem as well .

PR #985, opened 2022-09-06
Solution to Two-Pointers-2

meghadusane97:


MergeTwoSortedArray.java
```diff
@@ -0,0 +1,29 @@
+// Time Complexity : O(1)
```
jeelpatel18, line 1: time is m+n and space is constant

RemoveDuplicates.java
```diff
@@ -0,0 +1,29 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
jeelpatel18, line 2: Again it is constant

jeelpatel18 (commented):
Good job!!!

jeelpatel18:
Good job!!!

PR #1023, opened 2022-10-13
Completed Sprint Two-Pointers-2: Please Review

vedantnib:


mergeSortedArray.py
```diff
@@ -0,0 +1,14 @@
+#Time Complexity: O((n+m)log(n+m))
```
osandeepreddy, line 1: Please write optimized solution also

osandeepreddy:
Please write optimized solution also

PR #1032, opened 2022-10-24
Two pointers-2: Please review

pavansai1:


p2.java
```diff
@@ -0,0 +1,32 @@
+//TC: O(n)
```
daredevilteja, line 1: TC is O(m+n)

daredevilteja:
Good Work! Keep it up

PR #1080, opened 2023-01-06
Completed Two-Pointers-2 - Please review

ShridharSattur1:


MergeSortedArray.java
```diff
@@ -0,0 +1,54 @@
+// Time complexity : O(n)
```
daredevilteja, line 1: TC is wrong. Please recheck it.

RemoveDuplicates.java
```diff
@@ -0,0 +1,21 @@
+public class RemoveDuplicates {
```
daredevilteja, line 1: Please mention TC and SC

daredevilteja (changes requested):

daredevilteja:
Please make requested changes.

PR #1157, opened 2023-06-09
Sprint Completed TwoPointer2 - Please Review

PreeGarg:


Problem2.py
```diff
@@ -0,0 +1,39 @@
+'''
+Problem: Merge Sorted Arrays
+Time Complexity: O(n), where n is given elements
```
ishika2741, line 3: Kindly check TC. You have two arrays here and the size of the two arrays is different on which you are iterating.

ishika2741 (commented):
Well done!

ishika2741:
Well done!

PR #1173, opened 2023-06-19
Two pointed 2

salonimodi:


ishika2741 (commented):
Good work! Kindly add TC and SC for all the problems.

ishika2741:
Good work! Kindly add TC and SC for both the problems.

PR #1210, opened 2023-08-13
update two pointers -2 

rajashekarcs2023:


two pointers 2
```diff
@@ -0,0 +1,17 @@
+Intuition : Using two pointers, move slow only when count <= k, Fast will moving always as in for loop
+
```
ishika2741, line 2: Kindly add TC and SC as well.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1257, opened 2023-10-16
Added solutions to the code

nithind-sudo:


ishika2741 (commented):
Kindly submit third problem as well. Also add Tc and Sc for all the problems.

ishika2741:
Kindly submit third problem as well. Also add Tc and Sc for all the problems.

PR #1361, opened 2024-04-21
Two pointers -2

saiashok:


vyankateshmisar:
Good job Sai.

vyankateshmisar:
Good job Sai.

PR #1403, opened 2024-07-05
completed two-pointer-2

dhikki-study:
#########88. Merge Sorted Array##############################################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I was taking the pointers in beginning but it was messing up my list when trying to append than post class understood that pointer should be taken from last to maintain order

// Your code here along with comments explaining your approach in three sentences only
We had 3 pointers p1 at the end of list1 element p2 at end of list1 itself and p3 at end of list 2, now we compare p1 and p3 and which ever is largest we place it in p2 and move the pointers and list will be sorted

        

########80. Remove Duplicates from Sorted Array II###########################################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I was not able to crack the problem, but after class tried it myself

// Your code here along with comments explaining your approach in three sentences only
We have fast and slow pointer, the fast pointer is to iterate and slow pointer is to collect
     

########240. Search a 2D Matrix II###################################################
// Time Complexity : class solution - m+n, my solution - nlogm
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I figured out using for loops to select the cell where element may be present and than applied BS but in class a better approach with TC m+n was taught

// Your code here along with comments explaining your approach in three sentences only
#in my way I ran for loop to identify the column where records are present and than on those columns I ran BS
# in class it was taught to start at 1 corner where on 1 side value is increasing and other side value is decreasing and keep traversing till we find the element

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,138 @@
+#########88. Merge Sorted Array############################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I was taking the pointers in beginning but it was messing up my list when trying to append than post class understood that pointer should be taken from last to maintain order
+
+// Your code here along with comments explaining your approach in three sentences only
+We had 3 pointers p1 at the end of list1 element p2 at end of list1 itself and p3 at end of list 2, now we compare p1 and p3 and which ever is largest we place it in p2 and move the pointers and list will be sorted
+
+
+class Solution:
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1433, opened 2024-08-11
solved

nehayj100:


Nandha2Tech (commented):
Great Work! TC for search a 2d matrix would be O(m+n).

Nandha2Tech:
Great Work! TC for search a 2d matrix would be O(m+n).

PR #1538, opened 2025-02-07
python

divyanshi-srivastava:


problem3.py
```diff
@@ -0,0 +1,20 @@
+'''
+// Time Complexity : O(m + n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Three line explanation of solution in plain english : same implementation as in class
```
sudhasr, line 5: Kindly do not mention as "same implementation as in class". The intension for asking you to describe the approach in 3 lines is to give you a practice to communicate the approach in a concise manner. Also, the TAs checking your solution might not know what was discussed in the class. We would want to understand the thought process behind your approach by reading it.

sudhasr (commented):
Great Job on all the solutions. Please do mention the 3 line explanation of your approach.

sudhasr:
Great Job on all the solutions. Please do mention the 3 line explanation of your approach.

PR #1563, opened 2025-02-11
two pointers 2

karthikdevuluri123:


Erlybird:
Good job! Please mention the TC and SC

Erlybird:
Good job! Please mention the TC and SC

PR #1617, opened 2025-05-30
3 problems complete

arihantjain2198:
3 problems complete

### DIFF
--- a/SearchIn2DMatrix.java
+++ b/SearchIn2DMatrix.java
@@ -1,0 +1,20 @@
+class SearchIn2DMatrix {
+    public boolean searchMatrix(int[][] matrix, int target) {
+        int n = matrix.length;
+        int m = matrix[0].length;
+        int r=0;
+        int c = m-1;
+
+        while(r<n && c>=0){
+            if(matrix[r][c]==target) return true;
+            else if(matrix[r][c]<target){
+                r++;
+            }else{
+                c--;
+            }
+        }
+
+        return false;
+        
+    }
+}
--- a/MergeSortedArray.java
+++ b/MergeSortedArray.java
@@ -1,0 +1,26 @@
+class MergeSortedArray {
+    //Complexity time : O(m+n);
+    // space : O(1)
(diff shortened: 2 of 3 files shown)


SearchIn2DMatrix.java
```diff
@@ -0,0 +1,20 @@
+class SearchIn2DMatrix {
```
sudhasr, line 1: what is the Time and Space complexities?

sudhasr (commented):
Good solutions, pls mention Time and Space complexity Analysis for all the solutions.

sudhasr:
Good solutions, pls mention Time and Space complexity Analysis for all the solutions.

super30admin:
1. Correctness:
   - MergeSortedArray: The solution correctly merges two sorted arrays in-place using a two-pointer approach from the end. It handles edge cases where one array is empty.
   - RemoveDuplicateInSortedArray: The solution correctly removes duplicates in a sorted array, keeping at most 2 duplicates (k=2). It maintains the relative order of elements.
   - SearchIn2DMatrix: The solution correctly searches for a target in a 2D matrix with rows and columns sorted in ascending order using a two-pointer approach.

2. Time Complexity:
   - MergeSortedArray: O(m + n) as correctly noted by the student.
   - RemoveDuplicateInSortedArray: O(n) as correctly noted by the student.
   - SearchIn2DMatrix: O(n + m) where n is the number of rows and m is the number of columns.

3. Space Complexity:
   - All solutions use O(1) additional space, which is optimal for these problems.

4. Code Quality:
   - The code is generally well-structured and readable.
   - Variable names are meaningful (p1, p2, r for pointers in MergeSortedArray; slow, count in RemoveDuplicateInSortedArray).
   - Minor improvement: In MergeSortedArray, the else if(p2<0) condition could be simplified to just else since the return statement is not necessary (the loop would naturally terminate).
   - In RemoveDuplicateInSortedArray, the variable k could be made a constant or passed as a parameter for more flexibility.

5. Efficiency:
   - All solutions are efficient and use optimal approaches.
   - No significant optimizations are needed, but the code could be slightly more concise in places.

PR #1669, opened 2025-07-08
Done Problem-2 and Problem-3 Two-Pointers-2

Shreya1093:
Problem-1 is yet to be done in class


super30admin:
- The student has done an excellent job in implementing the solution. The code is clean, efficient, and well-commented.
- One minor improvement could be to remove the edge case check for nums1 being null or empty, as the problem constraints guarantee that nums1.length == m + n and m, n >= 0. However, this does not affect the correctness or efficiency of the solution.
- The comments are helpful, but the student could also mention why starting from the end of the arrays is beneficial (to avoid overwriting elements in nums1 that haven't been processed yet).

PR #1684, opened 2025-07-23
Completed Two-Pointers-2

shantanujoshi25:


super30admin:
- Problem_1.py: The logic for handling duplicates is incorrect. You should focus on ensuring that the count of duplicates does not exceed 2 and that the array is modified correctly. Consider revising the conditions and the way you update the count and pointers.
- Problem_2.py: The solution is close to correct but has a logical error when `i < 0`. Instead of breaking early, you should continue copying the remaining elements from `nums2` into `nums1`. This can be done by removing the `break` statement and letting the loop continue until `j >= 0`.
- Problem_3.py: This solution is well-implemented and efficient. Good job! The code is clean and easy to understand.

Overall, you have a good understanding of the problems, but there are some logical errors that need to be addressed. Pay close attention to edge cases and ensure your logic covers all scenarios.

PR #1702, opened 2025-09-05
Add solutions for array and matrix problems

ravithakur2k:
Added three new Python files: MergeSortedArray.py implements in-place merging of two sorted arrays; RemoveDups.py removes duplicates allowing at most two occurrences; Search2DMatrix.py searches for a target value in a 2D matrix. All solutions use efficient pointer-based approaches and are verified on Leetcode.


super30admin:
- Strengths:
  - The solution correctly implements the merging of two sorted arrays in-place, which is efficient and optimal.
  - The code is well-commented and easy to understand.
  - Variable names are meaningful and enhance readability.
- Areas for Improvement:
  - The student included solutions for other problems (`RemoveDups.py` and `Search2DMatrix.py`) which were not part of the evaluation. It would be better to focus on the problem at hand to avoid confusion.
  - The comment about time complexity could be more precise by specifying O(m + n) instead of just O(n), as the problem involves two arrays of sizes m and n.

PR #1709, opened 2025-09-06
Completed Two-Pointers-2

vidyasreekumar:


super30admin:
- The student's solution is excellent and correctly implements the two-pointer technique for merging sorted arrays.
- The comments are helpful and explain the approach clearly.
- The code is clean and follows best practices.
- One minor improvement could be to remove the redundant while loop for `p1 >= 0` since the remaining elements in `nums1` are already in place and don't need to be moved.

PR #1736, opened 2025-11-30
Completed Two Pointers - 2

shreyargh:


super30admin:
Strengths:
- The time and space complexity analysis is accurate and matches the optimal solution.

Areas for Improvement:
- In "MergeSortedArray.java", the comment "// 1: We use 3 pointers to maintain different positions in both arrays." could be more precise. There are only two arrays involved, and the third pointer is for the merged array.
- In "RemoveDuplicates.java", the comment could be expanded to explain why the slow pointer starts at 1 and why the count is initialized to 1, as this might not be immediately obvious to someone reading the code.
- In "SearchInA2DArray.java", the comment could clarify why starting from the top-right corner is optimal, as this is a key insight for the problem.

PR #1747, opened 2026-01-09
Done Two-Pointers-2

vaishnavi2231:


super30admin:
- Strengths:
  - The time and space complexity analysis is accurate and clearly stated.
- Areas for Improvement:
  - In the "merge-sorted-array" solution, the print statement (`print(p1,p2,r)`) is unnecessary and should be removed in production code.
  - The comment for the "remove-duplicates-from-sorted-array-ii" solution mentions "O(2n)" time complexity, which is technically correct but can be simplified to "O(n)" since constants are dropped in Big-O notation.
  - The student could consider adding edge case checks (e.g., empty arrays) for robustness, though the current solutions handle the given constraints well.

PR #1755, opened 2026-01-11
Done with two-pointers-2

Mukul585:


super30admin:
- The student's solution for Problem2.java is excellent. It correctly implements the two-pointer technique to merge the arrays in-place, which is the optimal approach for this problem.
- The comments provided are clear and help in understanding the logic, which is a good practice.
- The variable names (p1, p2, idx) are appropriate and make the code easy to follow.
- The solution handles edge cases well, such as when nums2 has remaining elements after nums1 is exhausted.

PR #1760, opened 2026-01-12
Completed Two-Pointers-2

rbhargav0104:


super30admin:
- The solution provided does not match the problem description. It seems you may have misunderstood the problem or submitted the wrong file. The problem requires merging two sorted arrays into one, not searching in a 2D matrix.
- For the problem you solved (search in a 2D matrix), your solution is correct and efficient. The approach of starting from the top-right corner and moving left or down based on the comparison with the target is optimal.
- The code is clean and well-commented, which is good practice. However, ensure you are solving the correct problem in future submissions.
- If you intended to solve the merge sorted arrays problem, you would need to implement a two-pointer approach starting from the end of both arrays to merge them in-place.

PR #1769, opened 2026-02-21
Completed Two-Pointers-2

sarvanibaru:


super30admin:
Your solution for the "Merging of 2 arrays" problem is excellent. It correctly implements the three-pointer technique from the end, which avoids overwriting elements in nums1 that haven't been processed yet. The time and space complexity are optimal. The code is clean and easy to understand.

However, note that you submitted solutions for multiple problems in the same response. For future submissions, ensure that you only provide the solution for the problem being evaluated. This helps in maintaining clarity and focus.

One minor suggestion: In the for loop, you break when p2 < 0. This is correct because if nums2 is exhausted, the rest of nums1 is already in place. Alternatively, you could also write a while loop for the pointer p, but your current approach is efficient.

PR #1771, opened 2026-02-21
Complete Two-Pointers-2

PrakarshKamal:


super30admin:
Your solution for the merge problem is excellent. It correctly implements the optimal algorithm with O(m + n) time and O(1) space. The code is clean and easy to understand. However, note that the problem statement specifically asks for a solution to the "Merging of 2 arrays" problem, and you have included two additional solutions for other problems in the same submission. While those solutions might be correct for their respective problems, it is important to submit only the relevant code for the problem being evaluated. In a real coding interview or platform, you should ensure that your submission contains only the solution for the problem at hand.

